PR #1, opened 2015-08-06
Refractoring, and fixing some issues

caiwan:
I have refractored and redesigned a few things, to use DALI messaging in other applications:
- The entire code was reformatted to meet PEP 8
- Some classes have reasonable name
	- `daliserver` -> `DaliServer`

**Command type and objects**
- Command object was separated from their base types (`command.py` and `commands.py`)
- Content (*data*) of commands are stored in tuples (`self._data`); since DALI defines various package lengths up to 4 bytes, and this kind of stuff has to be handled,
	- has `__len__` magic method to determine it's length in bytes. This required to make able to encapsulate them. 
	- has `pack` property to pack the contents of the objects into a byte stream (or bytes in py3 ) - previously it was done in `daliserver.send()` method
	- has _FORMAT_STRING private property, which returns the struct fromat string. 
	- the constructor makes tulpe from it's parameter list; due to backward compatibility.
	- the class has a `check_command_format` static method to verify the type of the given tulpe
		- has tulpe type, with length [2..4]
		- each value is between [0..255]
		- otherwise throws a `ValueError` exception
		
**DaliServer**
- has context manager, which manages (open, close) the connection to the server.
- creating a response from a byte stream, with a previously sent command was separated, due to handle encapsulated DALI commands in other messages


sde1000:
Thank you for working on this - I've been neglecting it recently!  I hope to have time to go over your changes thoroughly at the weekend.

Just to let you know where I'm going with this (in case you want to get there first):

* I plan to split the commands for different device types out into different files
* I'm going to add the commands for incandescent lamp controllers and LED lighting
* I'm going to add support for the parts of the standard that require 24-bit frames
* I want to finish the "object model" code, and then write a django-based web interface using it

caiwan:
In a nutshell what I was up to with this project is that I have a custom DALI controller, which is able to communicate over ethernet. It uses custom packet structures, but DALI commands are encapsulated into it. At this point came this project handy for me; I've modified the project to take care of this part.

sde1000:
That sounds really useful!  Is it something you can share?

caiwan:
Unfortunately, it's not, and I shouldn't tell any further details of it.

sde1000:
Ah, fair enough.

I'm using the Tridonic DALI-USB interfaces at the moment, but they are *really* expensive for what they are and it's annoying not having proper documentation for their USB protocol.  At some point I expect I'll roll my own, maybe as a USB device or possibly a Raspberry PI "hat".

caiwan:
To tell the truth, we're facing the same exact issue right now.

sde1000:
Just letting you know what I've done while processing this merge request.  I'm trying to maintain a sensible history, so I'm cherry-picking and merging commits together where this seems appropriate.

* Pulled commits cb4e733 and 152ae9b ("Gitignore for PYcharm") and squashed them together
* Cherry-picked commit a5dbd1f ("Reformatted the entire code to meet PEP 8 standard.")
* Pulled the next four commits (up to "Import fix") and squashed them together with a sensible commit message
* At this point I stopped for bugfixing, because the unit tests would not run - you had broken the "from_bytes" mechanism.  (I checked it was still broken in caiwan:master before doing this.)

sde1000:
Ok, I've now committed a fix for the bugs introduced in the commits up to 86778dc ("Import fix") and pushed to my master branch. I'm going to comment on the rest of the commits and cherry-pick where I can.

sde1000:
Ok, I've pushed my own edit to add a context manager to daliserver and rename it to DaliServer; I then cherry-picked your commit "Unpacking response was separated from send()".  I haven't had a chance to test this yet - I'm working in the wrong office.

sde1000:
I've cherry-picked 498e65c ("Refractoring the rest of the code") although I amended the commit message.  14cdfa6 looks like it's largely the same edits on a different branch, which is most likely what led to the merge error in 01ab863.

I think at this point my tree incorporates all of your changes apart from the `__unicode__` to `__str__` ones, and this would be a good point for you to rebase your tree.  You are likely to have to add `from __future__ import unicode_literals` to most of the files, or ensure that when you use .format() your format string is explicitly unicode - but that's a good idea anyway.

The unit tests pass (`python -m unittest discover` from the project root), although I'm not in a place where I can test that the scripts and the DaliServer interface code actually work at the moment!  I'd appreciate it if you could do that.

sde1000:
Before I add the commands for LED and incandescent lamp controllers, I'm thinking about moving things around a bit.  What do you think of the following structure?

* dali/
    * exceptions.py
    * frame.py - forward and backward frames and transactions
    * command.py - command registry, interface to command decoding
    * commands/ - commands and their responses
        * gear/ - commands to control gear
            * base.py - commands from part 102
            * fluorescent.py - commands from part 201
            * emergency.py - commands from part 202
            * incandescent.py - commands from part 205
            * led.py - commands from part 207
            * ... commands from other parts 203, 204, ...
        * device/ - commands to control devices and events from them
            * base.py - commands from part 103
            * ... commands from parts 301, 302...
    * interface/
        * daliserver.py - interface to https://github.com/onitake/daliserver
        * tridonic.py - driver for Tridonic DALI-USB device

The change you've introduced making command data be a variable-length tuple is a good one, but I'm thinking of taking it further and making "Frame" be a class.  According to the standard (part 101 section 7.4) forward frames may contain an arbitrary number of data bits, with 16- and 24-bit frames being defined in the standard, 20- and 32-bit frames being "reserved", and all other sizes of frame being considered "proprietary".  When transmitted, frames also have a priority from 5 ("default") to 1; I'm not sure yet whether this priority should be stored with the frame or passed separately to the transmitter (I'm leaning towards "separate" but I could be convinced either way).

There's also the concept of a "transaction", which is a sequence of frames where the first frame is transmitted at priority 2, 3, 4 or 5 and all the remaining frames are transmitted at priority 1.  It would be useful to be able to pass a transaction to an interface as the "thing to be transmitted".

caiwan:
The file structure is fine. 

My request would be the same, keep the command - and responses - to be encapsulated for further use. I have a custom built controller which comunicates through a (virtual) serial port, I'm thinking on implementing that.

I'll take care of my brach.

sde1000:
Ok, I'll work on that next time I have a chance.  I'll close this pull request for now; do open another one if you have more to add.  (I'd be surprised if I didn't introduce any bugs in the command-line utilities when I updated them, and I haven't had a chance to test yet.)

PR #6, opened 2015-09-14
Dev

caiwan:
I have a few things to review 
- I had to put a `pythonpath` realigment to the unittests, to make it run with Pycharm's unittest runner. It works with `python -m unittest discover` too. 
- I've put a dummy TCP listener for debugging purposes, nothing special
- I've added gitignore rules to ignore Eclipse's project files too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,7 @@
 *.pyc
 *~
+.project
+.pydevproject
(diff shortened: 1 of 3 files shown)


sde1000:
Ok, all merged.

PR #7, opened 2015-09-18
Update license, update ignores

### LINKED ISSUE #4: Package License
Hi,

would you mind license this package under LGPL, MIT, APACHE or similar license?

Regards

rnixx:
fixes #4

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,12 @@
 *.pyc
 *~
-.project
-.pydevproject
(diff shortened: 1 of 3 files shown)


sde1000:
The GLPLv3 doesn't replace the GPLv3, it supplements it - so the licence file should be added to, not replaced.

Can you make the .gitignore changes be a separate commit?

PR #11, opened 2015-09-18
Fix and sort imports for test scripts

rnixx:


sde1000:
I cherry-picked this commit instead of merging, to keep the history simple.

PR #14, opened 2015-09-18
Remove superfluous blanklines, absolute imports all over the place, m…

rnixx:
…inor PEP8 issues, use commands module instead of command module in dali.bus

### DIFF
--- a/dali/__init__.py
+++ b/dali/__init__.py
@@ -1,1 +1,1 @@
-__all__=["address","command","interface"]
+__all__ = ["address", "command", "interface"]
--- a/dali/__init__.py
+++ b/dali/__init__.py
@@ -1,1 +1,1 @@
-__all__ = ["address", "command", "interface"]
+#
--- a/dali/address.py
+++ b/dali/address.py
@@ -1,6 +1,5 @@
 class AddressTracker(type):
-    """Metaclass keeping track of all the types of Address we understand.
-    """
+    """Metaclass keeping track of all the types of Address we understand."""
 
     def __init__(cls, name, bases, attrs):
         if not hasattr(cls, '_addrtypes'):
@@ -10,8 +9,7 @@ class AddressTracker(type):
 
 
 class Address(object):
-    """An address for one or more ballasts.
-    """
+    """An address for one or more ballasts."""
     __metaclass__ = AddressTracker
 
     @classmethod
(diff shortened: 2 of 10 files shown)


sde1000:
That's quite a large diff! Did you use a tool, or do it all manually?

docstring style: I use emacs python-mode and it tends to want to put a blank line before the closing `"""` at the end of multi-line docstrings.  Does PEP 8 have anything to say about this?  Is there a way to turn it off?

sde1000:
I think the `__all__` in `dali/__init__.py` is out-of-date; if there's no good reason for it we should remove it, probably just leaving a docstring.

sde1000:
I notice that PEP 257 (referenced from PEP 8) says that single-line docstrings should have the closing `"""` on the same line.

rnixx:
I'm using eclipse with pydev which tells me potential issues, but all changes are done manually.

rnixx:
ok, PEP 257 is not in yet, will have a look

sde1000:
`dali/command.py` - `_check_destination`: first line of docstring should be a summary of the function.  "Destination" should start with lower-case 'd' because it is a parameter name.

sde1000:
Could you separate this into two commits, please (I know it's annoying!) - one with formatting changes only, and the other with bugfixes only (fixing 'command' -> 'commands')?

rnixx:
i do not know anything about emacs python mode, nor do i think PEP 8 prevents you to add a blankline. I'm pretty sure you can turn the blankline check off somehow...

Most (if not all) projects i'm working with have no blanklines at the end of a docstring, so removal on cleanup happens somwhat self-acting ;)

rnixx:
Separating the PR. Really? the code in bus.py never worked as it was and i wonder if there is a real benefit when having two commits?

sde1000:
I can do it if you like - I can pull your commit as-is, fix it up, and make two separate commits myself.

rnixx:
yes please +1

sde1000:
Ok, will do!

rnixx:
thank you

sde1000:
Caught a new bug introduced by reformatting: you changed "r.value != True" to "r.value is True" in bus.py - this is one reason I prefer formatting fixes to be entirely in their own commit, because it's clearly wrong.  If bugfixes and formatting fixes are in the same commit then it's not immediately apparent that this is wrong.

rnixx:
ok, mea culpa, sorry

sde1000:
Ok, I pulled your changes, separated out the bus.py bugfix, fixed the bug introduced by reformatting, and squashed the formatting changes all down to one commit using git rebase --interactive.  I've pushed it to my master branch.

Issue #3, opened 2015-08-19
CommandTracking is completely broken


caiwan:
Having a metaclass to keep track of all class types, is not working. We'll have to find an other way to do it.

sde1000:
In what way is it broken?

caiwan:
The _commands property doesn't seem to exist, unless you intantitate these classes. I didn't found any solution, rather than registering these command types by hand.

sde1000:
Are you using python 3?  The way metaclasses are declared has changed between python 2 and python 3:

Python 2:

```
class foo(object):
    __metaclass__ = bar
    pass
```

Python 3:

```
class foo(metaclass=bar):
    pass
```

caiwan:
Is there any chance to make it work with version (at the same time)?

sde1000:
It's a bit of a pain; you either write it in quite a convoluted way, or make use of a compatibility library like [six](https://pythonhosted.org/six/).

How important is python 3 compatibility to you at the moment?  I think I mentioned before that I do intend to port this to python 3 at some point, but keeping python 2 compatibility is important to me for now; I make use of this library in some pubs that use [quicktill](https://github.com/sde1000/quicktill) which is python 2 for now.  (Although, having said that, it might be time for me to look at converting quicktill to python 3 now most of its library dependencies have been ported.)

caiwan:
I'll rollback my changes to see if it's broken under py 2.7 or not. I'll also get know that if py3 is really nesesarry, or not. (I'm fine w/ py2, but the environment, where it will be used... )

caiwan:
What exactly does the `_devicetype` parameter in the commands, and how it sohuld really work?

sde1000:
IEC 62386 defines many different types of control gear.  Some commands are common between all of them; these are defined in IEC 62386-102.  Other commands are specific to particular types of control gear, for example fluorescent ballasts, emergency lighting controllers, LED lighting controllers, and so on and these commands are defined in IEC 62386-201, 202, 203, and so on.  Part 201 defines commands for device type 0, part 202 defines commands for device type 1, and so on.

These device-specific commands (referred to in the standard as "application extended commands") all occupy the same "space" - they are encoded in ways that conflict with each other.  To issue one of these commands, you first send the EnableDeviceType(n) command with the appropriate value of n, so only devices of that type will respond to the command.

[...]

PR #19, opened 2015-09-27
Use Python 3 compatible decorator for setting metaclasses

rnixx:
Referring to https://github.com/sde1000/python-dali/issues/3: @caiwan may you give py3 another try?

@sde1000 the compat code is actually taken from six, no need to depend to the whole library in this case IMHO.


sde1000:
Merged as a cherry-pick, after merging your pep8 cleanup.

PR #21, opened 2015-09-29
UART/Serial communication

caiwan:
- Common `DriverInterface` to create custom DALI drivers with it. 
- Communicates with a custom DALI controller through UART
- Binrary protocol, But can be extended with 7-bit ASCII
- Using frames, which starts with one byte length, and ends with one byte XOR CRC. 
- CRC check on the recieve side is missing yet.

### DIFF
--- a/dali/driver/daliserver.py
+++ b/dali/driver/daliserver.py
@@ -67,7 +67,7 @@ class DaliServer(DriverInterface):
             if not self._s:
                 s.close()
 
-        response = self.unpack_response(command, result)
+        response = self._unpack_response(command, result)
 
         if response:
             logging.info(u"  -> {0}".format(response))
--- a/dali/driver/daliserver.py
+++ b/dali/driver/daliserver.py
@@ -1,0 +1,77 @@
+"""
+Daliserver
+Communicates with daliserver through tcp
+"""
+
+__author__ = 'caiwan'
+
+from __future__ import print_function
+import socket
+import logging
+from dali.command import Command
+
+from dali.interface import DriverInterface
+
+
+class DaliServer(DriverInterface):
+    """
+    Communicate with daliserver
+    (https://github.com/onitake/daliserver)
+
+    NB this requires daliserver commit
+    90e34a0cd2945dc7a15681f11647e708f858521e or later.
+
+    """
+
+    def __init__(self, host="localhost", port=55825,
+                 multiple_frames_per_connection=False):
+        self._target = (host, port)
+        self._s = None
+        self._multiple_frames_per_connection = multiple_frames_per_connection
+
+    def __enter__(self):
+        if self._multiple_frames_per_connection:
+            self._s = socket.create_connection(self._target)
+        return self
+
+    def __exit__(self, *vpass):
+        # print(vpass)
+        if self._multiple_frames_per_connection:
+            self._s.close()
+            self._s = None
+
+    def send(self, command):
+        if self._s:
+            s = self._s
+        else:
+            s = socket.create_connection(self._target)
+
+        assert isinstance(command, Command)
+        message = command.pack
+
+        logging.info(u"command: {}{}".format(
+            command, " (twice)" if command._isconfig else ""))
+
+        # Set a default result which may be used if the first send fails
+        result = "\x02\xff\x00\x00"
+
+        try:
+            s.send(message)
+            result = s.recv(4)
+            if command._isconfig:
+                s.send(message)
+                result = s.recv(4)
+        except:
+            raise
+        finally:
+            if not self._s:
+                s.close()
+
+        response = self.unpack_response(command, result)
+
+        if response:
+            logging.info(u"  -> {0}".format(response))
+
+        return response
+
+
--- a/dali/driver/uart.py
+++ b/dali/driver/uart.py
@@ -1,0 +1,79 @@
+
+"""
+UART communication driver for DALI
+"""
+
+import logging
+import serial
+from dali.command import Command
+from dali.interface import DriverInterface, ParameterError
+
+
+class DaliUART(DriverInterface):
+    """
+    Uart driver for DALI
+    """
+    
+    def __init__(self, com, baud=115200):
+        """
+        :param:com port name
+        :param:baud baud rate - 11520 default
+        """
+        self._s = None
+        self._uart = (com, baud)
+
+    def __enter__(self):
+        """
+        Entry point of the context manager 
+        """
+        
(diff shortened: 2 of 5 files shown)


dali/driver/daliserver.py
```diff
@@ -0,0 +1,77 @@
...
+        self._s = None
+        self._multiple_frames_per_connection = multiple_frames_per_connection
+
+    def __enter__(self):
+        if self._multiple_frames_per_connection:
+            self._s = socket.create_connection(self._target)
+        return self
+
+    def __exit__(self, *vpass):
+        # print(vpass)
```
rnixx: Please remove this commented print statement

dali/driver/uart.py
```diff
@@ -0,0 +1,109 @@
...
+"""
+
+import logging
+import serial
+import struct
+from dali.command import Command
+from dali.interface import DriverInterface, ParameterError
+
+
+class DaliUART(DriverInterface):
```
rnixx: Which UART device is meant here? May you give the driver a more concrete name?

dali/driver/uart.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        assert isinstance(command, Command)
+        assert self._s is not None
+        assert isinstance(self._s, serial.Serial)
+
+        recv_data = None
+        response = None
+
+        try:
+            #
```
rnixx: Please remove this comment

dali/interface.py
```diff
@@ -9,68 +7,27 @@ class CommunicationError(Exception):
...
-        if response:
-            logging.info(u"  -> {0}".format(response))
-
-        return response
+    def send(self):
+        """
+        Send command, and returns its response
+        :return:
+        """
+        raise RuntimeError("Not implemented method")
```
rnixx: Please raise NotImplemented instead of RuntimeError

dali/driver/uart.py
sde1000: This is a good idea.  For example, the serial driver I committed a few days ago (completely untested!) implements the Tridonic SCI2 protocol.  What's the name of the protocol yours implements?

dali/interface.py
```diff
@@ -9,68 +7,27 @@ class CommunicationError(Exception):
...
 
 
-class DaliServer(object):
-    """Communicate with daliserver
-    (https://github.com/onitake/daliserver)
+class ParameterError(Exception):
+    pass
 
-    NB this requires daliserver commit
-    90e34a0cd2945dc7a15681f11647e708f858521e or later.
```
rnixx: Did this hint went somewhere else? If not, please re-add to README or so. Further we might want to ask daliserver developers creating a tag, since a commit id is not that ideal

dali/interface.py
```diff
@@ -9,68 +7,27 @@ class CommunicationError(Exception):
...
+class DriverInterface(object):
```
rnixx: Should the class name either be Driver or Interface? DriverInterface is somewhat confusing, even InterfaceDriver seems more appropriate.

dali/interface.py
rnixx: Also you might consider abc module for defining abstract classes, e,g, https://github.com/wichert/lingua/blob/master/src/lingua/extractors/__init__.py#L136

dali/driver/uart.py
caiwan: It has no specific name, its a raw 8-bit protocoll. The message heads with a length, and trailing with the crc of the message. Thats it.

dali/interface.py
caiwan: I had no proper name for them, but I'd refractor if you have better suggestion.

dali/driver/uart.py
rnixx: > I do have a custom controller
* Self made? baptise it ;)
* If not, use product name

dali/interface.py
caiwan: I missed this, thanks.

dali/interface.py
rnixx: For me InterfaceDriver sounds most "naturally". Not that important though

dali/driver/uart.py
caiwan: It's sort-of self-made, but I can't release any part of it. I do believe that anyone can build a microcontroller stuff at home, and implement this. [Like this one.](http://siwawi.bauing.uni-kl.de/avr_projects/dali/)

I'll rename it to RAW, or something.

rnixx: Then maybe GenericUART, with a protocol description?

caiwan: Ok. Description is [here](https://github.com/caiwan/python-dali/blob/master/dali/driver/uart.py#L46)

sde1000:
These are all useful comments.  I suspect I'm not going to merge this as-is, though, because we don't know what a good hardware interface looks like yet; defining an abstract base class for one before we've figured this out will just slow down experimentation.

I think that, at least for now, I'm happy to merge code that drives various different bits of hardware; we can pick a place ("drivers/" for example) for it to live.  I don't want the drivers to depend on a common base class yet, and I'm not going to insist that they all share a common interface.

Some examples of what I'm getting at:

Drivers will need to work in systems that are single-threaded, multi-threaded, or event-driven.  What should the interfaces for these look like?  Do we want an async interface with callbacks, like [Twisted](https://twistedmatrix.com/trac/)?  Do we want a blocking interface as well, possibly built on top of it?  Or something else?

Some use-cases for this library, which will all want to use the device drivers in different ways:

1. Simple command-line utilities that send commands to DALI control gear
2. A DALI application controller that listens to events from DALI control devices and responds by sending commands to DALI control gear
3. A full web interface to a DALI system that covers tasks from commissioning (assigning short addresses to devices - a lengthy process that can tie up the network for a few minutes) to letting users change lighting preferences

The first will want a simple blocking interface.  The second will probably be best suited to an event-driven interface.  The third will probably want to be threaded, or at least exist as a worker thread servicing requests from other threads in a web server process.

caiwan:
At this point of the project, I think this structure is fine. We'll surely rewrite later. Although open an issue to discuss this.

rnixx:
@caiwan I share the objections regarding a common base class. If we can avoid rewrites for base API's in the first place we should do it.

@sde1000 as @caiwan said, we should open a new issue discussing the event-driven/threaded approaches. I personally want to use the lib in a hybrid approach (Twisted + Worker Threads)

caiwan:
I've createt another pull request to my feature branch instead of master [here](https://github.com/sde1000/python-dali/pull/22)

PR #22, opened 2015-10-01
Caiwan dev uart

caiwan:
I've cleaned up things, we discusssed [here](https://github.com/sde1000/python-dali/pull/21).


caiwan:
I dont relly get wy exactly wants it to pull unto master instead of `sde1000:caiwan-dev-uart`.

caiwan:
Better if we would working on driver interfaces to create and implement any kind of protocols for various devices.

PR #31, opened 2016-01-02
Fix Query device type response.

Boldie:
The problem was, the value from self.value contains an instance of a class,
which is not convertible to an int as it is needed for the dictionary to
look up the final type for returning


sde1000:
I've pushed a commit that fixes this without accessing the private _data attribute of the frame directly.  Let me know if it works for you!

PR #33, opened 2016-01-06
Command tracker fix + hash for frame

caiwan:
I've fixed command tracker and added a test case for them. Also added a hash method for Frames.

### DIFF
--- a/dali/frame.py
+++ b/dali/frame.py
@@ -156,6 +156,9 @@ class Frame(object):
         except:
             raise TypeError("Frame can only be added to another Frame")
 
+    def __hash__(self):
+        return hash(self._data)
+
     @property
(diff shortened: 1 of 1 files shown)


sde1000:
Hmm.  My original intent was for the list of commands to be kept in the Command class not the metaclass - then, if you need a method to access it, it should be a class method on Command.  (The metaclass is not intended to be part of the exposed API - in fact, it should probably be renamed with a '_' at the start of its name to make that clear.)

The hash method on Frame looks ok.

caiwan:
I was attempting to access the list of available commands from outside; due I was experimenting with 
I'd use `importlib` or something similar rather than messing with metaclasses.

PR #40, opened 2017-05-02
Consolidate code

rnixx:
- Add ``dali.exceptions`` module and move all DALI related exceptions to it.
- Move ``dali.interfaces`` to ``dali.drivers.daliserver``.
- Update Package structure documentation.


rnixx:
@sde1000 I guess this module can entirely be removed https://github.com/sde1000/python-dali/blob/master/dali/device.py ?

And should code from device package go to gear instead?

sde1000:
The code in `device.py` is used in `bus.py`.  Both of these modules are a work in progress.  They should go in a separate subpackage; as you point out, there's name clash between `device.py` and the `device` subpackage.  The new subpackage should probably be called `bus`.

Moving exceptions to a separate module is a good idea.  I think it's also worth adding an object hierarchy to them so that related groups of exceptions can be caught more easily.

The `device` subpackage is for DALI control devices as defined in IEC 62386 part 103.  The `gear` subpackage is for DALI control gear as defined in IEC 62386 part 102.  They should stay where they are.

rnixx:
Changes made as pointed out. I also moved tests to a dedicated package.

Whoever did the pycharm path hack, may you be so kind and try running the tests with this changes?

rnixx:
Thank you.

PR #43, opened 2018-01-04
Update general.py

Vince-Rock:
I think this should be 0xa9 instead of 0xaa.


sde1000:
I see you closed the pull request, but just making sure:

In my copy of the standard (BS EN 62386-102:2014) QUERY CONTROL GEAR FAILURE is definitely listed as 0xaa in table 15 on page 52.  Do you have another reference?

dgomes:
This pull request was very strange... the author deleted himself few minutes after issuing the pull request...

PR #47, opened 2018-10-02
Hasseb - Adopt new api

dgomes:
I rewrote the Hasseb driver according to the new API.

Implementing Async version is virtually impossible since there is no state information in the driver messages.

### DIFF
--- a/dali/driver/hasseb.py
+++ b/dali/driver/hasseb.py
@@ -1,3 +1,7 @@
+import logging
+import struct
+from time import sleep
+
 from dali.command import Command
 from dali.driver.base import AsyncDALIDriver
 from dali.driver.base import DALIDriver
@@ -8,13 +12,10 @@ from dali.frame import BackwardFrame
 from dali.frame import BackwardFrameError
 from dali.frame import ForwardFrame
 from dali.exceptions import ResponseError, MissingResponse
-from time import sleep
-import logging
-import struct
 
 
-DALI_USB_VENDOR = 0x04cc
-DALI_USB_PRODUCT = 0x0802
+HASSEB_USB_VENDOR = 0x04cc
+HASSEB_USB_PRODUCT = 0x0802
 
 HASSEB_DRIVER_NO_DATA_AVAILABLE = 0
 HASSEB_DRIVER_NO_ANSWER = 1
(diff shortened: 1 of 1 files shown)


rnixx:
@dgomes Thanks for your work!

> Implementing Async version is virtually impossible since there is no state information in the driver messages.

Have you tried starting the USB listener against the hasseb device and listen to the bus for incoming frames from e.g. a hardware switch? If this sends commands, I think you can at least expect the Hasseb device to send the response for a received command prior to send an arbitrary incoming forward frame if there is a timing collision. But that's just guessing. I don't own such a device. But what I've seen at a short research, the firmware is available from NXP. To be sure about the behavior, it's source needs to be digged.

dgomes:
I don't have such switches, but I will look into the source code an try an educated guess. I did this refractor after recently starting to get No answers responses from the hasseb controller. Do you own one?

rnixx:
> Do you own one?

unfortunately not

sde1000:
I'll merge this, but please can you squash it into one commit with a sensible descriptive commit message first?

dgomes:
Sure thing!

I will nonetheless try to address @rnixx comment first by looking into the firmware source code. Give me a couple of days.

dgomes:
Read the Windows App source code and the NXP SDK. 

This device is very basic :) no support what so ever for DALI switches... everything is Sync.

dgomes:
Good to go!

dali/driver/hasseb.py
```diff
@@ -1,128 +1,193 @@
...
+            return
+        elif isinstance(frame, BackwardFrame):
+            if self._pending:
+                command, callback, kw = self._pending
+                callback(command.response(frame), **kw)
+            else:
+                logger.error("Received frame for no pending command")
         else:
-            response = None
+            logger.error("Received frame is not BackwardFrame")
```
rnixx: Can you please add a comment that here's the place to handle incoming forward frames?

dali/driver/hasseb.py
```diff
@@ -1,128 +1,193 @@
...
+            HASSEB_USB_PRODUCT,
+            bus=bus,
+            port_numbers=port_numbers,
+            interface=interface
+        )
+
+    def send(self, command, callback=None, **kw):
+        data = self.construct(command)
+        if command.response is not None:
+            self._pending = command, callback, kw
```
rnixx: A more robust implementation would imo be a fifo queue for pending requests. This way, in theory some can send multiple requests at once. Not sure if hasseb firmware can handle this, needs to be checked. If firmware requires one request/response cycle to be finished before processing next one, at least an exception would be appropriate pointing out that theres an already pending request.

rnixx:
@dgomes Awesome! Thank you

dali/driver/hasseb.py
dgomes: Never dealt with a DALI install beyond mine. 

Is it possible to have more than one controller ? (Does it make sense for a controller to receive a ForwardFrame) ?

dali/driver/hasseb.py
dgomes: Firmware can't handle multiple requests. 

Implementing a queue would require a background thread to consume the queue elements or a true async implementation (using asyncio).

Throwing an exception is not an option, sometime commands fail with this controller and you would be locked (commands fail quite often with this controller, that why I added the retry algorithm with backoff)

dali/driver/hasseb.py
rnixx: Sure it's possible to have more than one controller. DALI is multi master. Everything which comes from even a simple DALI switch sends forward frames.

dgomes: It can't receive unsolicited frames no... multi master with Hasseb would become a big no no...

dali/driver/hasseb.py
dgomes: the firmware waits for the answer itself... if the answer hasn't arrived yet you receive the NoAnswer frame that I'm parsing already

dgomes:
Thank you soo much 👍

PR #60, opened 2020-01-19
New hasseb firmware, drivers, and graphical user interface

hasseb:


dali/bus.py
```diff
@@ -130,11 +128,93 @@ def assign_short_addresses(self):
...
+                group2 = i.send(gear.QueryGroupsEightToFifteen(sa))
+                try:
+                    group1 = group1.value.as_integer
+                    group2 = group2.value.as_integer
+                except:
+                    self._devices[sa].group = None
+                else:
+                    self._devices[sa].groups = self.parse_groups(group1, group2)
+
+    def parse_groups(self, group1, group2):
```
dgomes: This all function can be summarised as
```
return str([j for j in [group1 & 1 << i for i in range(7)] if j != 0])[1:-1] + str([j+8 for j in [group2 & 1 << i for i in range(7)] if j != 0])[1:-1]

File: dali/driver/hasseb.py [hunk 2]
```diff
@@ -13,181 +13,236 @@
 from dali.frame import ForwardFrame
 from dali.exceptions import ResponseError, MissingResponse
 
+from dali.gear.general import *
+
+from PyQt5.QtWidgets import QApplication
```
dgomes: IMHO we should not depend on Qt...

dali/driver/hasseb_firmware/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Updating the firmware
+
+The firmware of hasseb USB DALI Master can be updated through USB. By booting
+the device with jumper JP1 short circuited, the device appears as a mass
+storage to your computer. The firmware can be update by removing the old
+firmware.bin file from the mass storage and copying the new firmware to the
+device. The name of the firmware file does not need to be changed. The most
+resent firmware can be found in this folder.
+
```
dgomes: Can we revert to the old firmware by backing up the existing file ?

dali/driver/hasseb.py
hasseb: Qt is required only when using graphical user interface. If graphical user interface is not used the two QT lines can be commented out. Any suggestions how this can be done without manually commenting the two lines out.

dali/driver/hasseb_firmware/README.md
hasseb: Yes. Old firmware can be found from www.hasseb.fi/dali as well

dali/bus.py
hasseb: Might be. Not sure where the commas are added but need to test this. I'm not sure if this is the most self explaining way to do it.

dali/driver/hasseb_firmware/README.md
dgomes: Can you provide a changes list ?

I'm really excited about this :) but my setup has been working flawlessly for 2 years and don't won to mess with that...

dali/driver/hasseb.py
dgomes: Make it an anonymous callback, that is setup on initialisation

dali/driver/hasseb_firmware/README.md
hasseb: No I can't, there are too many fundamental changes. The new firmware should be fully documented in the firmware readme file.

sde1000: Under Linux, you can't reliably write the firmware by copying the new firmware to the device.  You need to access the block device directly.  See https://github.com/sde1000/hasseb-usb/blob/master/scripts/flash.sh for a script that does this (although you will need to edit the script to set the firmware filename).

dali/driver/hasseb.py
hasseb: I need to figure this out

dali/driver/hasseb_firmware/README.md
hasseb: You are right. I need to clarify this somehow.

dali/bus.py
```diff
@@ -130,11 +128,93 @@ def assign_short_addresses(self):
                     i.send(gear.ProgramShortAddress(new_addr))
                     r = i.send(gear.VerifyShortAddress(new_addr))
                     if r.value is not True:
-                        raise ProgramShortAddressFailure(new_addr)
+                        ProgramShortAddressFailure(new_addr)
+                        print(f"Error in programming short address {new_addr}")
```
dgomes: `print` is probably some leftover debug code ?

dali/command.py
```diff
@@ -64,7 +64,8 @@ def value(self):
         if self._value is None and self._expected:
             raise MissingResponse()
         if self._value and self._value.error and not self._error_acceptable:
-            raise ResponseError()
+            print(ResponseError())
```
dgomes: Debug ?

dali/bus.py
hasseb: original code used raise, which crashed the software. I replaced the raise with printf. It is not acceptable that the code crash if verifying short address fails

dali/command.py
hasseb: same here. I think it is not ok to crash the software if something goes wrong

dali/bus.py
sde1000: The idea is that the caller must be aware of the possibility of programming the short address failing, and must catch the exception if they wish to continue in that circumstance.

dgomes: We should check why it crashes, `print` inside a daemon software is not good option. 

Missing a `try ... except` around the call to `search_bus` ?

hasseb: I will try to find a way to handle the exception in QT code

sde1000:
Thank you for all your work on this!  The new firmware is a huge, huge improvement over the original firmware.

I have to choose which parts of this pull request to accept into the project — and before I can do that, I need to clear up what this project is all about, because I have never really been explicit about it before now.

I wrote python-dali so that people could make use of DALI hardware without having to possess a copy of IEC-62386.  When I found myself in possession of a building with DALI lighting, it is the library that I searched for and could not find.  I made use of the resources I _could_ find — in particular, [daliserver](https://github.com/onitake/daliserver) enabled me to make use of interface hardware I was able to buy at the time — but eventually I became frustrated, bought copies of some of the parts of IEC-62386, and tried to do a proper job.

python-dali is intended as a reference, and as a base for other projects.  It's not intended to be an application in and of itself.  I want it to have the smallest number of dependencies possible, so that it can be applied in the broadest variety of applications.  So, from this pull request:

* the driver and the description of the USB protocol to talk to the new hasseb firmware will definitely be pulled
* I also intend to write an async version of the driver that works using hidraw
* the driver's dependency on Qt has to go.  (It looks like a register/callback situation anyway.)

I don't intend to pull the GUI code.  I think it deserves to be its own separate project, which can have python-dali as a dependency.

hasseb:
Hi, 

Thank you for your message. I understand that you don't want to merge
the graphical user interface to your project. I will fix the issues from
the pull request review. Maybe I will then close the pull request and
make a new fork including only the wanted modifications. Maybe I can
also contribute straight to your project. 

There were only few modifications to your code. The biggest
modifications I made to the bus.py file. My fist idea was to include the
functionality to the hasseb driver, but decided to modity the bus.py
file instead. But let's see, I first need to solve the issues from the
review comments. 

I had to change the hid library as the library you used, did not work
with Windows. hasseb device is now using hidapi and a separate hid.dll
(compiled from open source) is required for windows. I also removed the
async part from the driver, because I used QT for asynchronous receive. 

BR,
Hans 

Stephen Early kirjoitti 21.1.2020 01:41:

> Thank you for all your work on this! The new firmware is a huge, huge improvement over the original firmware. 
> 
> I have to choose which parts of this pull request to accept into the project -- and before I can do that, I need to clear up what this project is all about, because I have never really been explicit about it before now. 
> 
> I wrote python-dali so that people could make use of DALI hardware without having to possess a copy of IEC-62386. When I found myself in possession of a building with DALI lighting, it is the library that I searched for and could not find. I made use of the resources I _could_ find -- in particular, daliserver [1] enabled me to make use of interface hardware I was able to buy at the time -- but eventually I became frustrated, bought copies of some of the parts of IEC-62386, and tried to do a proper job. 
> 
> python-dali is intended as a reference, and as a base for other projects. It's not intended to be an application in and of itself. I want it to have the smallest number of dependencies possible, so that it can be applied in the broadest variety of applications. So, from this pull request: 
> 
> * the driver and the description of the USB protocol to talk to the new hasseb firmware will definitely be pulled
> * I also intend to write an async version of the driver that works using hidraw
> * the driver's dependency on Qt has to go. (It looks like a register/callback situation anyway.)
> 
> I don't intend to pull the GUI code. I think it deserves to be its own separate project, which can have python-dali as a dependency. 
> 
> --
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub [2], or unsubscribe [3].
 

Links:
------
[1] https://github.com/onitake/daliserver
[2]
https://github.com/sde1000/python-dali/pull/60?email_source=notifications&amp;email_token=ACFLBFEMBM3NNWAVIJN764LQ6YZBVA5CNFSM4KI3MPLKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEJOBWMQ#issuecomment-576461618
[3]

dali/driver/hasseb.py
hasseb: All QT dependencies removed from hasseb.py driver

dali/driver/hasseb_firmware/README.md
hasseb: Instructions to update drivers in Linux added to readme.md

dali/bus.py
hasseb: bus.py reverted back to the original one. required functionality implemented in hasseb.py

dali/bus.py
dgomes: commas are added because we are casting the list into a str

dgomes:
The Gui dir should go :(

rnixx:
@hasseb please create a hasseb-ui package - or name it as you like - in your account containing the GUI code.

hasseb:
Yes, I will remove the gui folder and finalize the other comments hopefully very soon

rnixx:
@hasseb Another minor thing. I'm not sure whether python-dali is the right place for the device firmware binary and the protocol documentation either. IMO this should go to the hasseb related package as well. Anyway the hasseb_firmware folder must be moved out of the driver directory since this is a python package. Also, having the device firmware and docs in your own repo makes maintenance way easier since you can just handle it the way you want. @dgomes, @sde1000

PR #64, opened 2020-04-15
Finish removing python2 compatibility code

sde1000:
This gets rid of the compat module and all `__future__` imports.


dali/frame.py
```diff
@@ -1,23 +1,12 @@
...
 import struct
-import sys
-
-
-if sys.version_info < (3,):
-    _integer_types = (int, long)
-else:
-    _integer_types = (int,)
 
+_integer_types = (int,)
```
rnixx, line 3: This is used for checking for numeric values with isinstance and has no benefit anymore. better check isinstance(foo, int) directly.

sde1000: Fair enough!

PR #67, opened 2020-05-07
hasseb firmware and drivers without graphical user interface

hasseb:
I would like to merge hasseb firmware and drivers. All comments from the earlier pull request should be fixed. The gui folder was removed.


dgomes:
What's in the .dll file ?

sde1000:
I assume it's something to do with accessing HID devices under Windows.  It needs to be a dependency, not something included in the repo.

dali/bus.py
```diff
@@ -140,4 +140,4 @@ def assign_short_addresses(self):
                     i.send(gear.Terminate())
                     raise NoFreeAddress()
                 low = low + 1
-        i.send(gear.Terminate())
+        i.send(gear.Terminate())
```
sde1000, line 143: No need to change this file.

dali/command.py
```diff
@@ -278,4 +278,4 @@ def __str__(self):
             "{:02x}".format(c) for c in self._data.as_byte_sequence)
         return "({0}){1}".format(type(self), joined)
 
-from_frame = Command.from_frame
+from_frame = Command.from_frame
```
sde1000, line 281: No need to change this file.

dali/gear/general.py
```diff
@@ -279,6 +279,26 @@ class GoToLastActiveLevel(_StandardCommand):
...
+class ContinuousDown(_StandardCommand):
+    """Dim down using the set fade rate.
+
+    targetLevel shall be set to minLevel and a fade shall be started
+    using the set fade rate. The fade shall stop when minLevel is
+    reached.
+    """
+    _cmdval = 0x0c
+
+
```
sde1000, line 301: I can't find ContinuousUp and ContinuousDown in my copy of the standard; opcodes 0x0b and 0x0c aren't listed.  Do you have a reference for these?

Changes to the supported commands would need to be in a separate pull request.

dali/gear/general.py
```diff
@@ -611,15 +631,15 @@ class QueryContentDTR0(_StandardCommand):
...
-              4: "incandescent lamp dimmer",
-              5: "dc-controlled dimmer",
+              3: "Low voltage halogen lamp",
+              4: "Incandescent lamp dimmer",
+              5: "DC-controlled dimmer",
               6: "LED lamp",
-              254: "none / end",
-              255: "multiple"}
+              254: "None / end",
+              255: "Multiple"}
```
sde1000, line 642: No need to make these cosmetic changes.

examples/hasseb_tests.py
```diff
@@ -0,0 +1,39 @@
...
+        DALI_bus.assign_short_addresses()
+        print("Address | Groups | Device type")
+        for i in range(len(DALI_bus._devices)):
+            print(f"{DALI_bus._devices[i].address} | {DALI_bus._devices[i].groups} | {DALI_bus._devices[i].deviceType}")
+    else:
+        print(f"Invalid argument {sys.argv[1]}")
+
+# Print error if invalid argument
+else:
+    print("Give only one argument")
```
sde1000, line 39: This test script looks like it could apply to any driver, not just the hasseb one.

dali/driver/hasseb.py
```diff
@@ -34,160 +50,410 @@ def __repr__(self):
...
-                return BackwardFrame(data[1])
+                return BackwardFrame(data[5])
             elif response_status == HASSEB_DRIVER_INVALID_ANSWER:
                 # 3: "Invalid Answer"
                 return BackwardFrameError(255)
+            elif response_status == HASSEB_DRIVER_TOO_EARLY:
+                # 4: "Answer too early"
+                self.logger.debug("Answer too early")
+                return HassebDALIUSBAnswerTooEarly()
+            elif response_status == HASSEB_DRIVER_SNIFFER_BYTE:
```
sde1000, line 142: It looks like this ought to be an exception, rather than a 'magic' return value.

dali/driver/hasseb.py
```diff
@@ -34,160 +50,410 @@ def __repr__(self):
...
+
+    def wait_for_response(self):
+        """Wait for response message.
+        """
+        for i in range(200):
+            if not self._pending:
+                return
+            else:
+                self.receive()
+
```
sde1000, line 253: All the code below this line is device-independent and doesn't belong in a driver module.

hasseb: Yes, but there are two issues:
1. In the bus.py you do not recommend to use it
2. QT, which I'm using in the graphical user interface, don't like exceptions, so I had to remove exceptions.

examples/hasseb_tests.py
hasseb: Yes, it should work with any driver. I didn't find a very simple test example from examples folder so I wrote a one. Maybe renaming the file would work.

dali/gear/general.py
hasseb: I have the most resent DALI standard where the new commands are listed. I can send you a screenshot by email.

dali/driver/hasseb.py
sde1000: 1. You could put it in the GUI project.
2. If the driver raises exceptions, they should be caught by the GUI code (and suitable action taken) before they propagate up the call stack to any Qt code.

dali/gear/general.py
sde1000: That would be great, thank you.

dali/bus.py
hasseb: .dll is required in windows. The original HID library works only in Linux. Using hidapi.dll the code works both in Linux and in Windows. Maybe a drivers folder would be a better place.

hasseb: The .dll is compiled from here https://github.com/signal11/hidapi

hasseb: And in Linux its a dependency

dali/driver/hasseb.py
hasseb: Thanks I will solve this somehow, it is required only for the hasseb_test.py. I will use your bus.py in the test and make my own implementation to the gui folder.

hasseb: Or do you prefer removing the hasseb_test.py?

sde1000:
I'm going to put my test kit back together tomorrow and go through this.  Hopefully I can also update the async driver to work well with it.

hasseb:
There were also some discussion about the firmware folder in the earlier pull request. I think it is up to Stephen to decide will it be included or not.

dali/bus.py
sde1000: The code you linked to at https://github.com/signal11/hidapi is a C library with no Python wrapper.  I can't find a Python library corresponding to the hidapi calls you've written; both the common ones (https://github.com/trezor/cython-hidapi and https://github.com/jbaiter/hidapi-cffi) have different APIs.

hasseb: see the readme
https://github.com/hasseb/python-dali/tree/master/gui

sde1000: Oh, I see!  There's yet another one!

dali/driver/hasseb_firmware/README.md
```diff
@@ -0,0 +1,219 @@
...
+  * 0x02: Response data received
+  * 0x03: Invalid data
+  * 0x04: Answer too early
+  * 0x05: Sniffing byte
+  * 0x06: Sniffing byte error
+
+* Byte 4: Data length in bits
+
+* Byte 5-7: Data
+
```
sde1000, line 219: I'm a bit confused about this protocol.  I was expecting something like:
* host -> device: send *this*, I'm expecting a response
* device -> host: I sent it!
* device -> host: I got a response!

And for send-twice commands:
* host -> device: send *this* twice
* device -> host: I sent it!
* device -> host: I sent it again!

It looks like there's no reply to the host at all for commands that don't expect a response?

sde1000: By the way, is there a github project for the firmware where I can submit issues?  I need to say things like "the string descriptors are faulty" but it's not really on-topic here.

rnixx: @sde1000 @hasseb I already stated that IMHO firmware related binaries and documentation belongs to hasseb org , best inside an own project, not here. thanks

sde1000: I definitely don't want to include binaries, but I think there's a case for including documentation about how to drive the various interfaces.  I might want to include my notes on driving the Tridonic/Lunatone DALI USB interface, for example.

rnixx: sure, but i think such docs belongs to the doc strings in the related driver implementation.

dali/driver/hasseb.py
sde1000: My plan for removing `bus.py` and replacement code that's driver independent is in pull request #68.

hasseb:
I think it is now better to remove the firmware folder and hasseb_test.py and the bus related code from the driver. Then we can finish this pull request. Any issues found from the firmware can be commented in my master here https://github.com/hasseb/python-dali.
All removed files are still available at my master above. Only the new driver file hasseb.py and the new commands in command.py will remain.

dgomes:
@hasseb what everyone is advising is that a *new* GitHub repository is created to handle all this stuff: firmware, GUI, documentation.

If you need help feel free to reach-me

hasseb:
I would like to make the firmware source code open. The problem is that it is based on an example code from NXP. It is highly modified, but there is still something from NXP.

dgomes:
> I would like to make the firmware source code open. The problem is that it is based on an example code from NXP. It is highly modified, but there is still something from NXP.

You can release a patch to the NXP code instead of releasing the modified version. This is something I've seen in other projects, eg. https://github.com/Koenkk/Z-Stack-firmware/tree/master/coordinator/Z-Stack_Home_1.2

sde1000:
Ok.  Nearly there.  Two things to do before I can merge this:

1. The additional commands should be in a separate pull request.  Please add a note in their docstrings giving the version of the standard in which they were introduced.
2. Please rebase+squash on current master.  (You'll need to force-push after doing this.)

I still have comments to make on the new firmware, but I'll wait until it has its own issue tracker; this isn't the place.

Thank you!

dali/driver/hasseb.py
```diff
@@ -2,192 +2,263 @@
...
+from dali.exceptions import DuplicateDevice
+from dali.exceptions import NoFreeAddress
+from dali.exceptions import NotConnected
+from dali.exceptions import ProgramShortAddressFailure
 
+import dali.gear.general as gear
+<<<<<<< HEAD
+from dali.address import Short
+=======
+>>>>>>> 300173044abf09535d6f2c9950b811553d168d2d
```
rnixx, line 21: conflict

dali/driver/hasseb.py
```diff
@@ -2,192 +2,263 @@
...
 HASSEB_DRIVER_NO_ANSWER = 1
 HASSEB_DRIVER_OK = 2
 HASSEB_DRIVER_INVALID_ANSWER = 3
+HASSEB_DRIVER_TOO_EARLY = 4
+HASSEB_DRIVER_SNIFFER_BYTE = 5
+HASSEB_DRIVER_SNIFFER_BYTE_ERROR = 6
+<<<<<<< HEAD
+=======
+
+>>>>>>> 300173044abf09535d6f2c9950b811553d168d2d
```
rnixx, line 46: conflict

dali/driver/hasseb.py
```diff
@@ -2,192 +2,263 @@
...
         return 'NO DATA AVAILABLE'
 
     __str__ = __repr__
 
+<<<<<<< HEAD
 class HassebDALIUSBNoAnswer:
+=======
+
+class HassebDALIUSBNoAnswer(object):
+>>>>>>> 300173044abf09535d6f2c9950b811553d168d2d
```
rnixx, line 59: conflict

PR #72, opened 2020-05-11
New DALI commands Continuous Up and Continuous Down

hasseb:
New DALI commands Continuous Up and Continuous Down as in IEC 62386-103 Edition 1.1 2018-09


hasseb:
IEC 62386-103 Edition 1.1 is the control device standard. The same commands are found from the gear standard IEC 62386-103 Edition 2.1 as well.

Issue #70, opened 2020-05-09
Driver and device discovery


sde1000:
The collection of device drivers isn't really "plug-and-play" at the moment.  All the example code imports the author's favourite driver and needs to be edited before being run to work with the others.  This has caused problems for people picking up the library for the first time (eg. #42, #44) and in any case is a bit tedious!

We need some kind of driver registry and device discovery module, so that simple scripts that just work with a single DALI device can call to get a suitable driver with the API they need.  Where multiple devices are present, or the device being used isn't automatically discoverable, the device to use could be specified by environment variable or on the command line.  More complex code that works with multiple devices at once also needs a way of naming devices (eg. in its config file or on the command line), so we need a way to go from a string description of a driver/device to a driver instance.

ferdinandkeil:
By device you mean the DALI interface?

sde1000:
Sorry, yes, that's unclear!  Yes, I mean the interface between whatever the code making use of python-dali is running on, and the physical DALI bus.

ferdinandkeil:
I think this should be solved by streamlining the driver API. At least for me, most of the confusion came from the huge differences in how the drivers are implemented.
Also, this sounds to me like you plan to come up with a general CLI for python-dali and that would probably better off as its own project.

Issue #69, opened 2020-05-09
Tidying up drivers


sde1000:
Comments originally from @rnixx in #68:

Hi Stephen,

> I still don't think there's a "winner" amongst the several driver APIs 

I don't think we should talk about a "winning" API. Actually we have 2 approaches. The "old style" implementation in ``dali.driver.base``, with straight forward sync API and async implementation with callbacks. And the asyncio based implementation which uses the "modern" way of life. Both have their value and usecases. And finally there's the daliserver module which needs to be refactored to one kind of a driver or another, or both.

What we have in the driver package is some kind of a mess. Here are some thoughts from my side:

* We should make clear that we have asyncio and non asyncio drivers.
* We should share code among these drivers if possible. This primary includes the device specific frame extraction and construction code.
* We may share backend specific code if it makes sense. In any case we have asyncio style backends and old style backends.
* The hid module should be split up inside an asyncio package. We have a tridonic package where the non asyncio driver lives and the asyncio variant lives in hid.py. Same for hasseb driver. On the other hand the hid class is actually the hid device related asyncio backend.

This is what the hid module may be split up to:

    driver
        asyncio
            backend.py
                hid
            tridonic.py
                tridonic
            hasseb.py
                hasseb
            ...

I always felt you do not like the driver API I wrote very much. Maybe I'm wrong, either way I'm totally fine with this. I am not able to switch over to asyncio based main loop anytime soon, and I think other people also will or cannot use asyncio for one reason or another. Therefor we should just cleanup the package and document what's there and why.

rnixx:
Thanks for opening a dedicated issue. As low hanging fruit and for the most self documenting value we should start with splitting up the hid module into an asyncio subpackage. Are you fine with this? I can help on this one if you like.

sde1000:
> I don't think we should talk about a "winning" API.

Indeed.  I don't want to mandate a particular API at all; I see the drivers as an "optional extra" to the library, rather than its core; I want the library to be useful in environments where these drivers aren't appropriate at all, for example running on MicroPython in an embedded controller.

All the library code outside the drivers package needs to work whatever concurrency model the caller is using.

> straight forward sync API and async implementation with callbacks

[...]

PR #74, opened 2020-05-15
Quaddali feature

ferdinandkeil:
This is the driver for my custom DALI interface. The design will be published as open-source.

The DALI to serial conversion is pretty basic, mainly just translating what happens on the DALI line to serial data. Therefore some timing is implemented in the driver. The interface supports listening to a DALI bus as well, but that is not implemented in the driver yet.

I have so far tested the driver on a bus with a single device. I will do some testing with multiple devices and longer cable runs in the future.


dali/driver/base.py
```diff
@@ -235,3 +237,69 @@ def close(self):
...
+        while timeout and abs(start-perf_counter()) < timeout:
+            data = self._serial.read(1)
+            if len(data) > 0:
+                break
+        # DALI backframes are max. 1 Byte long, so that is all that is needed
+        if len(data) <= 1:
+            return data
+        else:
+            raise ValueError(f'read method returned too many bytes ({len(data)})')
+
```
sde1000, line 292: This depends on the details of the device protocol.  For example, the [Tridonic serial<->DALI interface](https://www.tridonic.com/com/en/download/data_sheets/DALI_RS232_Interface_en.pdf) sends 3 bytes back in response to every command.

ferdinandkeil:
@sde1000 I have added `pyserial` as a requirement to the `setup.py`, hoping that this PR will now pass the CI tests. Please let me know if there any other changes I need to make.

sde1000:
There's still an issue with dependencies.  You unconditionally import `serial` in the driver.base module, which means that (for example) someone using the `driver-unipi` extra and not the `driver-quaddali` extra would still need to have `pyserial` installed.

There's still a fair bit of work to be done tidying up driver interfaces, as discussed in #69 and #70.  It might be best to merge this now and sort it out later as part of doing that work.  I wonder what @rnixx thinks?

ferdinandkeil:
> There's still an issue with dependencies. You unconditionally import `serial` in the driver.base module, which means that (for example) someone using the `driver-unipi` extra and not the `driver-quaddali` extra would still need to have `pyserial` installed.
It was my understanding that it would be down to the user to check for the requirements before importing a driver. From the top of my head I can't think of an elegant way of conditionally importing `serial` while also letting the user know that the driver will now be available. If you have any ideas please let me know.

Would the CI test pass now? Or do I need to change the Travis configuration as well?

> There's still a fair bit of work to be done tidying up driver interfaces, as discussed in #69 and #70. It might be best to merge this now and sort it out later as part of doing that work. I wonder what @rnixx thinks?
As said in #70 I'd suggest starting work on a command line tool that builds on top of `python-dali` as a reference implementation. Maybe I'll find time to dabble with this later this year.

sde1000:
The CI test does pass now, because it ignores extras and drivers.

dali/driver/base.py
```diff
@@ -234,3 +236,66 @@ def close(self):
         self._disconnecting = True
         self._stop_listening.set()
         super(USBListener, self).close()
+
+class SerialBackend(Backend):
+    """``Backend`` implementation for Quad-DALI-USB-Interface. Uses serial connection."""
```
rnixx, line 241: Please update the docs, backend is not Quad-DALI bound.

dali/driver/base.py
```diff
@@ -234,3 +236,66 @@ def close(self):
...
+        @param timeout: read timeout in seconds; set to 0 or None to disable
+        @return data read from the interface (bytes)"""
+
+        # internal read timeout is set to 10 ms to increase responsiveness
+        self._serial.timeout = 0.010
+        # to compensate for the short read timeout and to use the timeout argument, the read is repeated until
+        # more time than specified has elapsed
+        start = perf_counter()
+        data = bytes()
+        while timeout and abs(start-perf_counter()) < timeout:
```
rnixx, line 283: You can define the read timeout when creating the serial.Serial instance.

dali/driver/base.py
```diff
@@ -1,5 +1,7 @@
 import threading
 import usb
+import serial
```
rnixx, line 3: This import and the usb import should be conditional.

setup.py
```diff
@@ -24,6 +24,7 @@ def readme():
     ],
     extras_require={
         "driver-unipi": ["pyusb", "pymodbus"],
+        "driver-quaddali": ["pyserial"],
```
rnixx, line 27: I'd prefer naming the extra requirements after the installed dependencies.

    extras_require={
        'usb': ['pyusb'],
        'serial': ['pyserial'],
        'modbus': ['pymodbus']
    }

rnixx:
>  It might be best to merge this now and sort it out later as part of doing that work.

I think extra requirements and conditional imports should be fixed before merging as this are low hanging fruits. @ferdinandkeil don't hesitate if you need help.

dali/driver/base.py
ferdinandkeil: That would be a deviation from ``Backend``.

I also think it is useful to be able to specify the ``timeout`` on a per-read basis. Of course, this could also be done using a public ``timeout`` attribute.

dali/driver/base.py
ferdinandkeil: I would therefore add a ``try ... except`` around the import. Do you have any suggestions on what to do when the import fails? Do some logging? Raise a warning?

setup.py
ferdinandkeil: Consider it done.

dali/driver/base.py
ferdinandkeil: 👍

dali/driver/base.py
rnixx: I have not taken a look at the serial implementation. Maybe it's possible to override the timeout on the serial instance.

rnixx: https://github.com/pyserial/pyserial/blob/master/serial/serialutil.py#L359

it is possible to set the timeout on an instance. However, setting it causes _reconfigure_port to be called which looks like a heavy operation at a first glance.

 https://github.com/pyserial/pyserial/blob/master/serial/serialposix.py#L358

maybe a timeout can be held at the backend instance and set only if it changed.

ferdinandkeil: In my implementation the ``timeout`` of the serial connection is fixed and the actual timeout is handled by the backend code. I think this was done to make the code more responsive and alleviate issues with lost bytes.

I will move the assignment to the ``__init__`` method, so it's only done once.

rnixx: I added a note in #69. There is the question if we really need cycle based timeouts.

dali/driver/base.py
rnixx: My favourite pattern would be an inline import

    class Foo:
        def __init__(self):
            try:
                import serial
            except ImportError:
                raise RuntimeError('Foo requires pyserial but it is not installed')
            self.conn = serial.Serial()

rnixx: as alternative you can use a marker

    try:
        import serial
        SERIAL_INSTALLED = True
    except ImportError:
        SERIAL_INSTALLED = False

    class Foo:
        def __init__(self):
            if not SERIAL_INSTALLED:
                raise RuntimeError('Foo requires pyserial but it is not installed')

dali/driver/base.py
```diff
@@ -266,16 +273,17 @@ def __init__(self, port, baudrate=115200, bytesize=serial.EIGHTBITS, parity=seri
         # flush all buffers to remove old data before checking the port
         self._reset_output_buffer()
         sleep(0.050) # just a precaution if the previous command sent some data
-        self._reset_input_buffer()
+        self._reset_input_buffer(
```
rnixx, line 276: You lost a closing bracket

dali/driver/base.py
```diff
@@ -238,7 +241,7 @@ def close(self):
         super(USBListener, self).close()
 
 class SerialBackend(Backend):
-    """``Backend`` implementation for Quad-DALI-USB-Interface. Uses serial connection."""
+    """``Backend`` implementation to communicate with DALI interfaces over a serial connection."""
 
     def __init__(self, port, baudrate=115200, bytesize=serial.EIGHTBITS, parity=serial.PARITY_NONE,
```
ferdinandkeil, line 246: There's also an issue here, as I am using constants from ``serial`` in the constructor. This breaks with the import being in the constructor. Can't think of an easy fix, though.

ferdinandkeil: I am talking about line 243/246, hope that's clear. Haven't used this commenting feature before, so please bear with me.

rnixx: You can set the defaults to None and use the serial constants at init time:

    def __init__(self, bytesize=None):
        try:
            import serial
            bytesize = bytesize if bytesize is not None else serial.EIGHTBITS
        except ImportError:
            raise

ferdinandkeil: Thank you for that super elegant solution 👍

PR #78, opened 2020-08-11
Cleaning Hasseb driver

dgomes:
Finally got sometime to upgrade my setup with Hasseb controller.

The driver has these orphan imports

Also of note:
- It requires pyhidapi

From a previous PR from @hasseb :

"To install pyhidapi:  
git clone https://github.com/awelkie/pyhidapi.git  
cd pyhidapi  
sudo python3 setup.py install"

This is important as this library is not available as a pip package... can we replace this by something more standard @hasseb ?

### DIFF
--- a/dali/driver/hasseb.py
+++ b/dali/driver/hasseb.py
@@ -7,12 +7,6 @@ from dali.driver.base import DALIDriver
 from dali.driver.base import SyncDALIDriver
 from dali.frame import BackwardFrame
 from dali.frame import BackwardFrameError
-from dali.exceptions import BadDevice
-from dali.exceptions import DeviceAlreadyBound
-from dali.exceptions import DuplicateDevice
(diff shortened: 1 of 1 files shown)


dgomes:
Propose https://github.com/apmorton/pyhidapi

hasseb:
I will try if I get it work with both Windows and Linux, it may also work with Mac which would be nice as well.

dgomes:
@sde1000 could you possibly make a 0.7.1 release to include this fix ?

kbacia7:
@hasseb Hello, can you somewhere on wiki add information about update firmware for your devices? 
I mean, move somewhere to website [this](https://github.com/hasseb/python-dali/tree/3dbf4af3b3770431e7351057ea328b4dbcc3a355/dali/driver/hasseb_firmware)

PR #79, opened 2020-09-02
Bump release to include hasseb fixes

dgomes:
Last accepted commit is very important for installs using Hasseb driver. We need a new pip release.


sde1000:
Ok, ok!

dgomes:
Sorry... and Thanks :)

PR #80, opened 2020-09-09
Udev rules

jbaptperez:
Enhances the _udev_ rules:

* In addition to physical connection, allows support of logical _udev_ triggering using `udevadm` command,
* Does not remove any previously created symbolic link for the device anymore.


sde1000:
Thank you.

PR #86, opened 2021-04-19
Fix two small bugs in the asyncio Tridonic driver.

awahlig:
This fixes two bugs that I spotted while working with the Tridonic driver:

- ```asyncio.wait_for()``` was called in ```tridonic._bus_watch()``` without handling the ```TimeoutError```
- ```tridonic._outstanding``` dictionary was not being reset correctly in ```tridonic._shutdown_device()```


sde1000:
Goot catches, thank you.

PR #92, opened 2021-10-27
Support for DALI memory banks — continued

sde1000:
This pull request incorporates and builds upon all the commits from #76 by @ferdinandkeil.

For a full list of changes see the commits, but briefly:

* Style changes to match the rest of the library: declaring memory values as classes in the same way the rest of the library declares commands as classes
* Implementation of memory bank 0 from IEC 62386-102
* The ability to write memory values as well as read them, where the bus unit supports this
* Support for TMASK and MASK when reading from the memory banks defined by DiiA
* Unit testing using "real" (made-up) data rather than just memory bank default values

Code that works with @ferdinandkeil's pull request #76 should continue to work with this, although you may see slightly different types being returned for some memory values (eg. Decimal rather than float for scaled values) and there's the possibility of a `dali.memory.location.FlagValue` being returned for memory locations that have invalid contents or which have values TMASK or MASK.

I don't think this is finished yet, but I'm ready for comments from people testing the code. Outstanding issues include:

* I'm aware that the specifications for banks 0 and 1 changed in a non-backwards-compatible way between the 2009 and 2014 versions of IEC 62386. I've implemented a "legacy" version of bank 0 for now, but I think there has to be a neater way of handling this. Are we likely to see further incompatible changes between the 2014 version and later versions?
* I need to write more test cases, especially of the code paths for handling errors and invalid values
* I'm struggling to interpret the DiiA specification "DALI Part 251" regarding strings. It says "ascii character string" with range for each byte of [0, 0xFF]. However, ASCII is a 7-bit character set and there are no valid ASCII characters in the range [0x80, 0xFF]. What should I do if I encounter any? (If I use python's built-in "ascii" codec I will just get a UnicodeDecodeError exception.)


sde1000:
With the "latch while reading whole memory bank" feature and the tidy-up of exceptions, I think this is now ready for merging. Testing and comments would be appreciated!

PR #96, opened 2021-12-30
Update find_ballasts.py

cgommel:
Did not work because of old library syntax.


sde1000:
Thank you.

In practice you would want to use `dali.sequences.Commissioning()` for this job.

PR #97, opened 2022-01-25
Run driver tests

mhemeryck:
Was just checking changes to my fork and the upstream one and noticed that the CI setup changed.

Would it be OK to also run the driver tests I did create at the time?

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -29,3 +29,7 @@ jobs:
       - name: Test with pytest
         run: |
           pytest dali/tests
+      - name: Run driver tests
+        run: |
+          pip install pymodbus pyusb
+          pytest dali/driver/tests


sde1000:
I had a few reasons for leaving out driver tests by default. (And also leaving out the flake8 step.)

1. The drivers aren't part of the supported API of the library (at the moment; I do have a plan, I'm just being very slow)
2. Not all drivers have tests; only one of them does so far!
3. Driver tests have additional dependencies, whereas the main set of unit tests only depend on the Python standard library

These are not necessarily all good reasons not to run them. Running them as an additional step, as your commit does, should be fine but may need looking at again in the future.

mhemeryck:
Hey, thanks for your reply!

I wouldn't necessarily mind if would leave these tests out for now. I was intending on looking into this library again since I wanted to augment my DALI-lighting setup. Since I noticed there were some changes since the last time I had looked at your library and since I did notice that the tests didn't run as part of the default test suite, I figured to at least run those again -- and I was glad to notice they still ran fine :slightly_smiling_face: 

Feel free to close this PR for now. If I find the time, I might work a bit further on the unipi driver in another PR.

PR #101, opened 2022-05-31
Add "level" support to fakes.py

sl-wallace:
This feature branch adds the necessary message handling to fakes.py to
handle the various DALI level control messages, e.g. DAPC,
RecallMaxLevel, etc. Such a feature was found to be helpful when
testing out higher level code which uses this library, by enabling the
use of fake DALI devices which correctly respond to certain messages.

Perhaps this isn’t the ideal place for such code to live, but since
fakes.py already implemented the majority of the necessary features
this seems to be the best fit within the existing structure.

Tests have been added to cover all of the additional code.

### DIFF
--- a/dali/tests/fakes.py
+++ b/dali/tests/fakes.py
@@ -102,6 +102,9 @@ class Gear:
                  devicetypes=[], random_preload=[],
                  memory_banks=(FakeBank0, FakeBank1)):
         self.shortaddr = shortaddr
+        self.level = 0
+        self.level_max = 254
+        self.level_min = 1
         self.scenes = [255] * 16
         self.groups = set(groups)
         self.devicetypes = devicetypes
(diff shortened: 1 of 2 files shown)


sde1000:
Looks good to me, and will be useful to include.

The only thing blocking me merging it is the use of bare `assert` in the tests — you should be using the various `assertSomething` methods of `unittest.TestCase` instead, eg. `self.assertEqual(resp.value, 0)` instead of `assert resp.value == 0`. See [the unittest documentation](https://docs.python.org/3/library/unittest.html) for details.

sl-wallace:
Thanks for that @sde1000 , I'll update my tests and push some more changes in the next day or two :smile:

sl-wallace:
Hi @sde1000, I've updated the test code to use `assertEqual` rather than bare `assert` statements :smile:

sde1000:
Thank you!

Issue #95, opened 2021-11-27
Schedule for dropping support for Python 3.6


sde1000:
At some point it would be useful to drop support for Python 3.6, for two main reasons:

1. So we can make use of features introduced in Python 3.7.
2. To make support for async code easier. (Some asyncio calls required in Python 3.6 are deprecated in Python 3.8 and removed in later versions)

But when?  Security support for Python 3.6 ends in December 2021, five years after its initial release; there's an argument that says support can be dropped after that. However, some distributions (eg. Ubuntu 18.04) will continue supporting Python 3.6 for longer; support for Ubuntu 18.04 ends in April 2023, for example.

Current distribution Python versions are:
* Ubuntu 18.04: 3.6
* Ubuntu 20.04: 3.8
* Debian buster ("oldstable"): 3.7
* Debian bullseye ("stable"): 3.9
* RHEL 8: 3.6
* RHEL 8.2: 3.8
* Arch: 3.9 (but releases back to 3.4 are installable)

I would like to pick a date between 23rd December 2021 and end of April 2023!

dgomes:
In other projects the rule of thumb is support major distributions (Ubuntu, Debia, RHEL) stable/LTS versions

sde1000:
There is very little in the queue for release 0.10 so far, certainly nothing that current users of python 3.6 + release 0.9 would find useful. Continuing to support python 3.6 in future releases is becoming troublesome, especially for code that includes type annotations or makes use of asyncio.

Therefore I'm going to say that release 0.10 will drop support for python 3.6 and the minimum supported version will be python 3.7. It looks likely that release 0.10 will happen when the 24-bit support code from @sl-wallace is merged and stable.

PR #103, opened 2022-06-27
Add support for DALI 24-bit event messages

sl-wallace:
IEC 62386 part 103 defines "event" messages, which can be sent by control devices to indicate various inputs or changes. This commit introduces support for encoding and decoding these messages, as well as a handful of new sequences for making it easier to work with them.
    
Support has been added for IEC 62386 part 301 "push button" devices, further device types should be simple enough to add in later. Unsupported types will be safely ignored by the library.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -21,6 +21,10 @@ jobs:
           python -m pip install --upgrade pip
           pip install flake8 pytest
           if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+      - name: Install dependencies (Python 3.6)
+        if: ${{ matrix.python-version == '3.6' }}
+        run: |
+          pip install dataclasses
       - name: Lint with flake8
         run: |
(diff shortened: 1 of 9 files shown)


sl-wallace:
Ah. I was trying to use the handy `from __future__ import annotations` feature to help with type hinting in this event message code - but it fails because Python 3.6 doesn't support this and the automated checks flag this.

@sde1000 when do you expect #95 to have a set date for dropping Python 3.6 support? If it's soon then I might just wait a while for this pull request, but if it's not until 2023 then I'd better go and refactor or remove my type hints!

sl-wallace:
Apologies for all the spam with the commits and re-runs of the tests, I think I can avoid using the annotations import and still get it to work for Python 3.6. I'll close this PR and create a new branch which is a bit cleaner, without all the one-line commits.

sde1000:
I think this is the right time to retire support for python 3.6 — this PR (well, the subsequent one) is likely to be the bulk of release 0.10; it's a new feature so there are no existing users to worry about!

Issue #102, opened 2022-05-31
Managing DALI gear through abstractions


sl-wallace:
I'm currently working on building a DALI integration for Home Assistant. I've got some code so far which broadly works, although needs much more testing and optimisation before it could be considered usable.

I'll make the source for it available once I've got something a bit more stable, but I'm wondering what the best way to go about it is – at the moment I use `python-dali` as a separate library (with a few modifications, which I'm getting ready to submit as pull requests). On top of that I have classes which define what I've called "proxies". These proxies are initialised at start-up by scanning the DALI bus for gear, then each proxy keeps a copy of key pieces of data from the corresponding piece of DALI equipment (most importantly the level, but also things like serial number etc), which can then be queried on demand from any higher-level application. The proxies abstract away all of the DALI commands, and have simpler methods like `turn_on`, which takes care of the relevant commands. State information, such as level, is queried as needed over the DALI bus. I then have a separate Home Assistant component which uses these proxies to map into the necessary objects that Home Assistant uses.

My question is – does this "DALI system" code belong in this library? Or should I maintain it as a separate project? In following the Home Assistant guidelines, an integration should not have device-specific code itself, but should rather create a layer on top of a more general library. This makes sense, as it would mean the library I've created could be used as a basis for some other system without needing to remove the Home Assistant parts. I'd be happy to try and integrate my code into this project if it would be useful, but I can equally appreciate wanting to keep things separate.

sde1000:
(Will comment properly in about a week and a half - I'm running the bar at https://emfcamp.org/ this week and it's taking *all* my time!)

sl-wallace:
> I think you can use python-dali directly in the HA integration without any issues

Fair enough @dgomes – I'll take a look at your repo and perhaps do a similar thing.

[...]

PR #105, opened 2022-06-30
Add Lunatone LUBA RS232 driver

sl-wallace:
Lunatone make a series of DALI interface devices which communicate with a host over RS232 serial, using a protocol [defined by Lunatone](https://www.lunatone.com/wp-content/uploads/2021/04/LUBA_Protocol_EN.pdf) called 'LUBA'.

This pull request adds support into `python-dali` for LUBA interface devices, as briefly discussed in #102. The driver API largely follows the existing `hid` classes of drivers, including drop-in compatible support for sequences.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install flake8 pytest
+          pip install flake8 pytest pytest-asyncio pyserial-asyncio
           if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
       - name: Lint with flake8
         run: |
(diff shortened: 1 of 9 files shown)


sl-wallace:
I still need to add some better documentation for how to use the driver interface, I'll do that before I mark as "ready for review". I realise too that it's unlikely any other users of this library will have access to this specific Lunatone device (I'm [using this one](https://www.lunatone.com/en/product/dali-rs232-ps240ma/), although there are others), so testing by others will be hard to do. You'll have to take my word for it that it works! :grinning:

sl-wallace:
@sde1000 I've tidied things up now, and I think this is pretty much good to merge now. I expect I'll add some more robustness to it in the future, e.g. have it attempt to automatically reconnect after a failure, but for now it works well enough based on the testing I've done. Hopefully it might be useful for others too!

PR #106, opened 2022-08-11
Add DALI DT8 colour gear support

sl-wallace:
Based on [a fork I'd found](https://github.com/sde1000/python-dali/compare/master...petrilgner:python-dali:master) by [Petr Ilgner](https://github.com/petrilgner/) I've added support for IEC 62386-208 DALI-2 DT8 colour temperature control gear. I've kept his original commits by cherry-picking them, so that the correct attribution remains in the commit history.

As well as the `dali.gear.colour` module, there's a new `dali.gear.sequences` module, which contains `SetDT8ColourValueTc()` and `QueryDT8ColourValue()` sequences.

I only have access to colour temperature gear, so support isn't completed for other types (e.g. RGBWAF, x-y coordinate). These should be easy enough to add in by someone who has such hardware to test with. All command opcodes have been implemented, even though I am not able to test all of them. (Note that opcodes 239, 244, 253 and 254 are "reserved" by the standard).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,6 +88,8 @@ Library structure
 
   - ``gear`` - DALI control gear as defined in IEC 62386; stable
 
+    - ``colour`` - Commands from part 209 (Device Type 8)
+
     - ``emergency`` - Commands from part 202
 
     - ``general`` - Commands from part 102
--- a/dali/gear/__init__.py
+++ b/dali/gear/__init__.py
@@ -7,4 +7,5 @@ extensions and control gear specific features).
 import dali.gear.general
 import dali.gear.emergency
 import dali.gear.incandescent
+import dali.gear.colour
 import dali.gear.led  # noqa: F401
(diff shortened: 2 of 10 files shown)


sde1000:
Although I see there are still a couple of TODO items (`QueryRBGWAFControl` and `QueryAssignedColour`) I'm happy to merge this.

sl-wallace:
OK, thanks for that. I'll take a quick look at sorting out some of the TODO items to see how much work it would be and if it's possible to meaningfully implement it without hardware - then submit and mark as ready.

sde1000 (approved):

sl-wallace:
I'm happy with this now, I haven't been able to test the `QueryAssignedColour` and `QueryRBGWAFControl` messages, but I have tests for them and they seem like they ought to work with appropriate control gear. If anyone uses it and it doesn't work then they'll just have to file a bug report! :grinning:

PR #108, opened 2022-10-04
Add support for DALI Occupancy Sensor devices

sl-wallace:
Implements support for commands, responses and events from IEC 62386 part 303: "Input devices — Occupancy sensor"


sl-wallace:
Hi @sde1000 , I've had a go at implementing Occupancy Sensor support. It seems to work well with the sensor that I have, although so far I just have the one to try it with.

Really the only thing that stuck out as a bit tricky was th
Screenshot from 2022-10-04 16-58-08
e way the events are encoded - there's multiple properties per event notification.

sde1000:
All good, thank you.

PR #115, opened 2023-03-03
two improvements for the `led` module

jktjkt:
- led: better printing of Part 207's Fast Fade Time

- led: one more bit for OPERATING MODE
  
  My copy of 62386-207:2009 says that command 252 (“QUERY OPERATING MODE”) also has bit#4 defined as "non-logarithmic dimming curve active".

### DIFF
--- a/dali/gear/led.py
+++ b/dali/gear/led.py
@@ -207,13 +207,30 @@ class QueryOperatingMode(_LEDCommand):
     _cmdval = 0xfc
 
 
+class FastFadeTimeResponse(command.NumericResponse):
+    """
+    Response to "Min Fast Fade Time" and "Fast Fade Time".
+
+    Refer to Part 207 section 9.13 and Table 1 "Fast fade time".
+    """
+
+    def __str__(self):
+        if isinstance(self.value, int):
+            if self.value == 0:
+                return "shortest"
+            if self.value > 27:
+                return f"out of range ({self.value})"
+            return f"{self.value * 25} ms"
+        return self.value
+
+
 class QueryFastFadeTime(_LEDCommand):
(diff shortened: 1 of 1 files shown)


jktjkt:
For the "OPERATING MODE" commit I'd appreciate some guidance on choosing an approach which plays best with the other parts of the library. I tried to look at the int-based enumerations, but the actual values did not appear to be used anywhere, so in the end I chose a simpler approach.

jktjkt:
Ah, right, indeed. Let me drop that patch, and fix the linter error in the other one.

sde1000:
`NumericResponse` subclass implementation note: I'd access the `value` attribute to get the value, and be prepared to deal with the case where `value` is a string owing to missing response or framing error.

jktjkt:
> Fast fade time: shouldn't you be multiplying the response by 25 rather than 50 to get the fade time in ms?

Indeed, sorry for that. My earlier, unpushed, approach with enums named `ms_0`, `ms_25`, `ms_50` etc was correct in this regard, but very verbose, and I apparently screwed up during refactoring.

> `NumericResponse` subclass implementation note: I'd access the `value` attribute to get the value, and be prepared to deal with the case where `value` is a string owing to missing response or framing error.

OK, that makes sense. I'll push an update later once I'm back to my ~~testing lab~~, I mean, home, and have a chance to verify how it works. Thanks for the quick review!

jktjkt:
Done. Tested on OTi DALI 80/220-240/24 4CHDT6/8 (for valid responses) and on Lunatone Jalousie relay (for a missing response).

jktjkt:
A gentle ping.

PR #117, opened 2023-04-01
DT8: support setting Tc limits, and other improvements

jktjkt:
I have a bunch of DT8 Osram OTi DALI DT6/8 devices which are by default set to something like 2700-6400K, but my LED strips have a different CT. Lunatone's SW won't allow me to change the *physical* Tc limits, so let's write a script which assists with this configuration.

Also some unrelated bugfixes as standalone commits.

### led: DT8: recognize Colour Control device type when parsing a BackwardFrame

Without this patch, dumping DALI traffic would show stuff like this:
```
QueryDeviceType(<address (control gear) 0>) -> multiple
QueryNextDeviceType(<address (control gear) 0>) -> LED lamp
QueryNextDeviceType(<address (control gear) 0>) -> BackwardFrame(8)
QueryNextDeviceType(<address (control gear) 0>) -> none / end
```
### led: DT8: configuration commands need sending twice

This is as per 62386-209:2011, section 11.3.4.2, "Application extended configuration commands".

### led: DT8: fix a copy/paste error in `SetDT8ColourValueTc`

Fixes: b706666 Expand DT8 support with tests and sequences

### led: DT8: a helper for conversion between K and Mirek for Tc

The standard talks about "Mirek", which is the SI lingo for Mired. I think that everybody else uses Kelvin, so let's add a helper function.

### examples: getting and setting a Tc for DT8 CCT LEDs

### DIFF
--- a/dali/gear/colour.py
+++ b/dali/gear/colour.py
@@ -185,6 +185,17 @@ class StoreXYCoordinatePrimaryN(_ColourCommand):
     _cmdval = 241
 
 
+class StoreColourTemperatureTcLimitDTR2(IntEnum):
+    """
+    Valid DTR2 values for the StoreColourTemperatureTcLimit command
+    """
+
+    TcCoolest = 0
+    TcWarmest = 1
+    TcPhysicalCoolest = 2
+    TcPhysicalWarmest = 3
+
+
 class StoreColourTemperatureTcLimit(_ColourCommand):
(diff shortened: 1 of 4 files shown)


sde1000:
All looks sensible, thank you. (I don't have a copy of the DT8 standard to check against, unfortunately.)

PR #118, opened 2023-04-23
tridonic: Fix bus traffic dumping

jktjkt:
I have more than one control device, which is why the `examples/async-dalitest.py` hits a framing error when checking if there are any control devices connected. When the bus_watch feature is enabled, this hit a Python error due to some incomplete rename/refactoring where this single occurrence was not changed by mistake.

Fixes: c72374c ("Update the async Tridonic DALI-USB driver to work with 24-bit frames")


sde1000:
Good catch. Thank you!

PR #119, opened 2023-04-30
hid: fix wrong variable name

jktjkt:
That's the endless joy of languages with no compile-time variable lookup.

Fixes: ea01ab3 ("asyncio-based drivers for USB devices that present as HID")


sde1000:
:-(

Too true.

The "drivers" and "examples" sections aren't don't have the same code quality standards as the main part of the library. That line had clearly never ever been tested...

PR #128, opened 2023-08-20
Control Devices: respect "resolution" when reading "inputValue"

jktjkt:
The standard (62386-103:2014, chapter 9.7) states that the "resolution" of an instance is not always 8-bits. If it's any longer, a proper read involves a transaction and a sequence of commands. If the number is not divisible by 8, extra trailing bits have to be truncated.

Tested on Lunatone DALI-2 MC (`86459532-2`) and Lunatone CS-2 (`86458670`). The push button coupler is a part-301 device where the standard says that the on-the-wire values are either `0x00` or `0xff`, but the actual resolution (and therefore the correct numeric value) is either 0 or 1. The other device, a combined movement sensor with a light meter, supports 11 bits of resolution for the lux meter and therefore it requires a `"QUERY INPUT VALUE"` followed by a `"QUERY INPUT VALUE LATCH"` and dropping the extra 5 bits.

### DIFF
--- a/dali/device/sequences.py
+++ b/dali/device/sequences.py
@@ -14,6 +14,9 @@ from dali.device.general import (
     DTR2,
     EventScheme,
     InstanceEventFilter,
+    QueryResolution,
+    QueryInputValue,
+    QueryInputValueLatch,
     QueryEventFilterH,
     QueryEventFilterL,
     QueryEventFilterM,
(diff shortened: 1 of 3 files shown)


sde1000:
Looks good, thank you.

Ideally there would be a test for this new sequence in [test_device_sequences.py](/sde1000/python-dali/blob/master/dali/tests/test_device_sequences.py)

jktjkt:
> Ideally there would be a test for this new sequence in [test_device_sequences.py](/sde1000/python-dali/blob/master/dali/tests/test_device_sequences.py)

Done -- and sorry for a "slight" delay :).

PR #135, opened 2024-03-03
    Add driver for Atx Led Dali Hat (AL-DALI-HAT).

rijimuge:
This uses the Raspberry PI on-board serial port to communicate at 19200 baud to the Dali Hat,
    the hardware on the hat adapts the UART serial data stream into DALI encoding.
    
    This uses pyserial to read from the dali hat device and implements a sync driver.

    An example using the Dali Hat driver is included, and an entry is added to the README file.

### DIFF
--- a/dali/driver/atxled.py
+++ b/dali/driver/atxled.py
@@ -39,7 +39,6 @@ class DaliHatSerialDriver(DALIDriver):
             self.LOG.exception("Could not open serial connection: %s", e)
             self.conn = None
 
-
     def read_line(self):
         """Read the next line from the buffer, refilling the buffer if necessary."""
         with self.lock:
@@ -153,16 +152,3 @@ class SyncDaliHatDriver(DaliHatSerialDriver, SyncDALIDriver):
             if command.is_query:
                 return command.response(resp)
             return resp
-
-
-if __name__ == "__main__":
-    """Usage: python atxled.py address value"""
-    from dali.gear.general import DAPC
-
-    logging.basicConfig(level=logging.DEBUG)
-    serial_port = "/dev/ttyS0"
-    dali_driver = SyncDaliHatDriver()
-    command = DAPC(int(sys.argv[1]), int(sys.argv[2]))
-    response = dali_driver.send(command)
-    print("DALI response:", response)
-    dali_driver.close()
--- a/dali/driver/atxled.py
+++ b/dali/driver/atxled.py
@@ -1,15 +1,10 @@
 from dali.command import Command
 from dali.driver.base import SyncDALIDriver, DALIDriver
-from dali.frame import ForwardFrame, BackwardFrame
+from dali.frame import BackwardFrame
 import logging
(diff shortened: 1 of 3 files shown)


dali/driver/atxled.py
```diff
@@ -0,0 +1,190 @@
+from dali.command import Command
+from dali.driver.base import SyncDALIDriver, DALIDriver
+from dali.frame import ForwardFrame, BackwardFrame
+import logging
+import sys
+
+# avoid import name collision with 'serial' driver
+original_sys_path = list(sys.path)
+sys.path = [p for p in sys.path if not p.endswith("python-dali/dali/driver")]
```
sde1000, line 9: This shouldn't be necessary. I suspect you're developing with `python-dali/dali/driver` as the CWD so you can run this module as `__main__`? Don't do that!

dali/driver/atxled.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def __init__(self, port="/dev/ttyS0", LOG=None):
+        """Initialize the serial connection to the DALI interface."""
+        self.port = port
+        self.lock = threading.RLock()
+        self.buffer = []
+        if not LOG:
+            self.LOG = logging.getLogger("AtxLedDaliDriver")
+            handler = logging.StreamHandler(sys.stdout)
+            self.LOG.addHandler(handler)
```
sde1000, line 31: If LOG is passed, you don't appear to be doing anything with it?

dali/driver/atxled.py
```diff
@@ -0,0 +1,190 @@
...
+                bytesize=serial.EIGHTBITS,
+                timeout=.2,
+            )
+        except Exception as e:
+            self.LOG.exception("Could not open serial connection: %s", e)
+            self.conn = None
+
+    def _read_line(self):
+        """Read a line from the serial connection."""
+        with self.lock:
```
sde1000, line 47: Would it be easier to use [`PySerial.read_until()`](http://map.emfcamp.org/) for this?

dali/driver/atxled.py
rijimuge: Good point-- I removed this and removed __main__ since it was only there ad probationem

dali/driver/atxled.py
rijimuge: Thanks! It will set self.LOG to the passed LOG now

dali/driver/atxled.py
rijimuge: Yes, that works very nicely!

rijimuge:
Hi, any additional thoughts on this? 

Want me to squash these commits and submit another pull request?

sde1000:
It's ok, I can squash. Sorry, I missed these updates. I'm on holiday this week and my development PC is turned off, I'll look at it next weekend.

sde1000:
Or maybe I can try github "squash and merge" which I've never used before...

PR #136, opened 2024-04-29
DeviceInstanceMapper: do not skip devices with inputDeviceError

jktjkt:
I have a bunch of non-pushbutton switches connected to some of my Lunatone DALI-2 MC pushbutton couplers. These devices unfortunately set the `"inputDeviceError"` bit whenever any of its inputs is closed/shorted for longer than `T_stuck`.

The 62386-301:2017 says (Table 4, page 15) that this timer is up to 255s, which is definitely not enough for a door sensor. The same standard then says that such errors should be propagated to bit 0 in the `"instanceErrorByte"`, which python-dali does not query. However, my devices unfortunately also propagate this status into the `"inputDeviceError"`, and that means that the initial discovery of instance numbers to instance types won't recognize these buttons at all.

I was not able to find a reference in the standard whether any `"instanceError"` propagates to the `"inputDeviceError"`, but even if this was an incorrect behavior, these devices exist, one can buy them, and presumably using them as a kind-of persistent contact inputs still makes sense; just to provide some data, I have:

- magnet and a reed switch for door status monitoring,
- a "permanent" switch for disabling automation at a hallway where our baby sometimes sleeps.

I think that both use cases make sense, and I would prefer python-dali to not skip over some input devices which actually work just fine.

Fixes: 0cfddb7

### DIFF
--- a/dali/device/helpers.py
+++ b/dali/device/helpers.py
@@ -127,8 +127,7 @@ class DeviceInstanceTypeMapper:
             if isinstance(rsp, QueryDeviceStatusResponse):
                 # Make sure the status is OK
                 if (
-                    rsp.input_device_error
-                    or rsp.short_address_is_mask
+                    rsp.short_address_is_mask
                     or rsp.reset_state
                 ):
                     continue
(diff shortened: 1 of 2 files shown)


sde1000:
Thank you.

PR #137, opened 2024-05-16
RFC: Add support for Lunatone DALI-2 USB dongle

plaes:
New hardware with some updates to protocol.

LUBA protocol:
https://www.lunatone.com/wp-content/uploads/2021/04/LUBA_Protocol_EN.pdf

DALI-2 USB device shows up as `/dev/ttyACM0` instead of previous `ttyUSB0`.

In datasheet rev 1.1 they added USB device support which features switchable power supply.
There's also small change in the `0x80` / `0x81` in rev 1.4...

As I don't have other devices, I'm unable to check for backwards compatibility, therefore this PR is marked as RFC...

### DIFF
--- a/dali/driver/serial.py
+++ b/dali/driver/serial.py
@@ -910,9 +910,29 @@ class DriverLubaRs232(DriverSerialBase):
                     f"Wrong event type 0x{self._buffer[1]:02x}, expected 0x21"
                 )
 
-            # QUERY DEVICE INFO supports two "sets" of data, this driver will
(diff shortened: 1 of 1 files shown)


sde1000:
I don't have any of the relevant hardware either, but it looks sensible to me! I'm happy to merge unless anyone who *does* have LUBA hardware drops in and says it causes problems.

sde1000:
Nobody has asked me not to, so I'm going to go ahead and merge this. Thank you!

PR #146, opened 2024-08-13
Improve HID driver support for hasseb

rohanramaswamy:
Rewrote the send and receive functions for hasseb in hid.py
Added support for firmware version reading for hasseb in hid.py
Updated  examples for flash and tc for hasseb tested on actual hardware


dali/driver/hid.py
```diff
@@ -714,40 +723,72 @@ def __init__(self, *args, **kwargs):
         self._response_available = asyncio.Event()
         self._response = None
 
+    def _initialise_device(self):
```
dgomes, line 726: when is this method called ?

dali/driver/hid.py
```diff
@@ -714,40 +723,72 @@ def __init__(self, *args, **kwargs):
...
+        transmitter_settling_time = 0
+        if command.sendtwice:
+            send_twice = 10 # 10 ms delay between messages
+        else:
+            send_twice = 0
+        frame = command.frame.as_byte_sequence
+        byte_a, byte_b = frame
+        data = struct.pack('BBBBBBBBBB', 0xAA, self._HASSEB_DALI_FRAME, self._sn,
+                           frame_length, expect_reply,
+                           transmitter_settling_time, send_twice,
```
dgomes, line 754: how is send_twice actually sent twice !?

dali/driver/hid.py
rohanramaswamy: Line 143 self._initialise_device(), it overrides hid class method.

dali/driver/hid.py
rohanramaswamy: It needs further testing to confirm, however in my testing the device sends it when send_twice is set > 0. value of send twice is used as delay in ms between the two sends.

dali/driver/hid.py
dgomes: please make a call to super()._initialise_device() and don't forget to set connected

dali/driver/hid.py
dgomes: you are creating a single packet, with a format which I haven't seen documented, can you link where you saw this struct ?

PR #149, opened 2025-05-16
emulating Lunatone's IoT protocol

jktjkt:
I wrote a small emulator which enables Lunatone's Windows GUI, the DALI Cockpit SW, to talk to an app which uses `python-dali`. As was suggested in #139, let's move this into upstream's git. Also, my goal is to make this usable from my application, so putting stuff into `examples/` (which is not packaged by default) wouldn't help me much. Let's stick this right below `dali/` unless there's some better place.

My USB-to-DALI interface is connected to a tiny headless Pi Zero, which is running some Python code (built on top of python-dali) for stuff like circadian color temperature control, automatic blinds, automatic light based on movement, etc. This all works well, but every now and then I have a need to run Lunatone's DALI Cockpit to perform some changes in my devices' configuration.

That's a Windows application, so I would usually fire up a VM with Windows and the DALI Cockpit, stop my automation, and then use usb-ip to export the USB device over network from my embedded server, e.g.:
```
ssh root@private.example.org systemctl restart usbip-bind@17b5\:0020.service
```
...then import that at my desktop, e.g.:
```
usbip attach -r private.example.org -b $(usbip list -p -r private.example.org | grep 17b5:0020 | cut '-d:' -f1 | awk '{print $1}')
```
...and finally attach that to the VM:
```
echo "<hostdev mode='subsystem' type='usb' managed='yes'><source><vendor id='0x17b5'/><product id='0x0020'/></source><address type='usb' bus='0' port='3'/></hostdev>" | virsh attach-device win10 /dev/stdin
```
That works, but it's a pain, and the automatic lights at the WC won't react to movement, and it's not great if it's dark in there.

So I got fed up and wrote this tiny emulator of the Lunatone DALI-2 IoT Gateway. The idea is that you can run this on the Linux machine which has the USB-to-DALI interface already connected, and you can let your Windows VM connect to that over IP, using the device's emulated Websocket entry point. It's possible to run this either as a standalone application, or you can simply put it into your application code.

### DIFF
--- a/dali/fake_lunatone_iot.py
+++ b/dali/fake_lunatone_iot.py
@@ -1,0 +1,216 @@
+import asyncio
+import dali.frame
+import dali.gear
+import dali.device
+from dali.device.helpers import DeviceInstanceTypeMapper
+from dali.driver.hid import tridonic
+from enum import Enum
+import json
+import logging
+from websockets.server import serve
+import websockets.exceptions
+
+"""
+Emulate the bare minimum of Lunatone's DALI-2 IoT Gateway over Websocket
+
+Sometimes it's useful to fire up Lunatone's DALI Cockpit, the proprietary
+Windows GUI application for addressing, configuring, managing, etc. a DALI
+installation. This code implements the bare minimum of their Websocket
+protocol, effectively emulating the "Lunatone DALI-2 IoT Gateway". If you
+run this code (or integrate the relevant two lines into your existing
+building automation system which uses python-dali internally), you can let
+the DALI Cockpit access the DALI bus without having to mess with, say,
+usb-ip and restarting services.
+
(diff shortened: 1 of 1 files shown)


sde1000:
I'm happy to take this, but I think it should be in `/examples/` and not `/dali/`.

I've pushed an update to the CI configuration that removes Python 3.7 and adds Python 3.12 — maybe move the file and rebase on that so the tests don't fail?

jktjkt:
> I'm happy to take this, but I think it should be in /examples/ and not /dali/.

I tend to agree, but then files below `examples/` are not packaged, and therefore I cannot take the code as-is and simply `import` than module into my application. What I'd like to prevent is having one copy in python-dali's `examples/` and another copy of the same file in my repo, if possible.

Is there a solution to that? If not, I'll respin this patch targeting `examples/` as you wish.

PR #152, opened 2025-11-21
Update Debian packaging

sikmir:
Fix error during debian package build:
```sh
ModuleNotFoundError: No module named 'serial_asyncio'
```


sde1000:
Thank you.

